PR #19, opened 2021-10-29
Fix TODOs

davidnevadoc:
Partially addresses #9

The more straightforward TODOs are solved but those that require further design cosdierations are left open


LukePearson1 (approved):
LGTM!

PR #21, opened 2021-11-04
Refactor Polynomial Commitment Scheme

### LINKED ISSUE #18: Replace `kzg10` with generic polynomial commitment scheme
Replace the `commitment_scheme` module that provides kzg10 commitments with `ark_poly_commit` to enable the use of other polynomial commitment schemes.

davidnevadoc:
The current implementation of KZG10 will be replaced by arkworks implementation.

Closes #18


src/circuit.rs
```diff
@@ -8,14 +8,18 @@
 
 use core::marker::PhantomData;
 
-use crate::commitment_scheme::kzg10::PublicParameters;
 use crate::constraint_system::StandardComposer;
 use crate::error::Error;
-use crate::proof_system::{Proof, Prover, ProverKey, Verifier, VerifierKey};
-use ark_ec::models::twisted_edwards_extended::{GroupAffine, GroupProjective};
+use crate::proof_system::{
+    Proof, Prover, ProverKey, Verifier, VerifierKey as PlonkVerifierKey,
```
LukePearson1, line 14: Is there a need to separate verifier key and plonk verifier key?

src/circuit.rs
```diff
@@ -305,18 +321,32 @@ where
...
     fn gen_proof(
         &mut self,
-        pub_params: &PublicParameters<E>,
+        u_params: &UniversalParams<E>,
         prover_key: ProverKey<E::Fr, P>,
         transcript_init: &'static [u8],
     ) -> Result<Proof<E, P>, Error> {
-        let (ck, _) = pub_params.trim(self.padded_circuit_size())?;
+        // XXX: KZG10 does not have a trim function so we use sonics and
+        // then do a transformation between sonic CommiterKey to KZG10
```
LukePearson1, line 329: Small nit `from` sonic may be better

LukePearson1 (approved):
Few very small nits. If addressed then feel free to merge

CPerezz (approved):
LGTM! Brilliant work!!!

src/circuit.rs
davidnevadoc: Right now we are overusing the name 'VerifierKey'. It refers to the commitments of arith polynomials, bool polynomials, etc. as well as the union of all of these keys used to verify the whole circuit.
With this change to the poly commit we have another VerifierKey so I decided to use an alias for the other one.
This problem is addressed in Issue #14

src/circuit.rs
davidnevadoc: I think it is not possible to implement a foreign trait between foreign structs. However it may be a good idea to create a wrapper for the commiter keys and implement the transformations there

PR #27, opened 2021-11-11
Add scalar conversion functions

### LINKED ISSUE #26: Add transformation functions between outer and embedded curve scalars
All variables in the circuit are viewed as the pairing friendly curve scalars regardless of their function (i.e. wire values, coordinates of an embedded curve point or embedded curve scalars).
It is necessary to implement safe transformations between these two types so that the embedded curve scalars can be used in scalar-point multiplications and as variables in the circuit.

davidnevadoc:
Resolves #26

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -80,7 +80,7 @@ pub fn get_domain_attrs<F: FftField>(
 /// Get a pairing friendly curve scalar `E::Fr` from a scalar of the embedded curve.
 /// Panics if the embedded scalar is greater than the modulus of the pairing firendly
 /// curve scalar field
-pub fn from_embedded_curve_scalar<
+pub(crate) fn from_embedded_curve_scalar<
     E: PairingEngine,
     P: TEModelParameters<BaseField = E::Fr>,
 >(
@@ -108,7 +108,7 @@ pub fn from_embedded_curve_scalar<
 /// Get a embedded curve scalar `P::ScalarField` from a scalar of the pariring friendly curve.
 /// Panics if the pairing frindly curve scalar is greater than the modulus of the embedded
 /// curve scalar field
-pub fn to_embedded_curve_scalar<
+pub(crate) fn to_embedded_curve_scalar<
     E: PairingEngine,
     P: TEModelParameters<BaseField = E::Fr>,
 >(
--- a/src/util.rs
+++ b/src/util.rs
@@ -5,7 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use ark_ec::{AffineCurve, ModelParameters, PairingEngine, TEModelParameters};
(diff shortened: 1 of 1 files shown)


davidnevadoc:
As far as I know these functions are only used in tests. For this reason they don't have their specific type of error and they are left in `util.rs`.

If this functionality is needed for something apart from tests and examples we should introduce types for each of these 2 scalars with their associated conversion functions and errors.

src/util.rs
```diff
@@ -77,17 +77,63 @@ pub fn get_domain_attrs<F: FftField>(
...
+    let modulus = <<E::Fr as PrimeField>::Params as FpParameters>::MODULUS;
+    if modulus.num_bits() >= scalar_repr.num_bits() {
+        let s = <<E::Fr as PrimeField>::BigInt as BigInteger>::from_bits_le(
+            &scalar_repr.to_bits_le(),
+        );
+        assert!( s < modulus,
+            "The embedded scalar exceeds the capacity representation of the outter curve scalar");
+    } else {
+        let m = <<P::ScalarField as PrimeField>::BigInt as BigInteger>::from_bits_le(
+            &modulus.to_bits_le(),
```
iquerejeta, line 99: Why are we doing this? if `modulus.num_bits() < scalar_repr.num_bits()`, then it should directly fail, right?

src/util.rs
```diff
@@ -77,17 +77,63 @@ pub fn get_domain_attrs<F: FftField>(
...
+    if modulus.num_bits() >= scalar_repr.num_bits() {
+        let s = <<P::ScalarField as PrimeField>::BigInt as BigInteger>::from_bits_le(
+            &scalar_repr.to_bits_le(),
+        );
+        assert!( s < modulus,
+            "The embedded scalar exceeds the capacity representation of the outter curve scalar");
+    } else {
+        let m = <<E::Fr as PrimeField>::BigInt as BigInteger>::from_bits_le(
+            &modulus.to_bits_le(),
+        );
```
iquerejeta, line 129: Same

iquerejeta (approved):
Looks good to me. However, if these functions are only used for tests, I would change the visibility of them. And, whenever there is the expectation of using these elsewhere, I would give a thorough description of which cases these functions must be used.

src/util.rs
davidnevadoc: I was thinking of the following case:
 1. The embedded scalar field  `P::ScalarField` is larger than the pairing friendly curve scalar field `E::Fq` and it needs more bits to be represented. 
 2. The used embedded scalar still 'fits' in `E::Fq` because it is lesser than its module.

iquerejeta: aha! I see. Then it makes sense. I (incorrectly) assumed that `into_repr()` was fitting only to the required bits. All good then 👍

PR #43, opened 2021-11-29
Add `big_arith_gate` and conditional point negate

tsunrise:
This PR adds the following gate: 
- `big_arith_gate` enforces `(a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI + q_o * c = 0` and returns `c`
- `big_arith` compute and enforces `c = (a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI`, and returns `c`
- `conditional_point_neg` returns  `-point` if `bit = 1` and `point` if `bit = 0`

Thanks for review!

### DIFF
--- a/src/constraint_system/arithmetic.rs
+++ b/src/constraint_system/arithmetic.rs
@@ -113,9 +113,7 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
     /// possible. Since it allows the end-user to setup all of the selector
     /// coefficients.
     ///
-    /// Forces `q_m * (w_l * w_r) + w_4 * q_4 + q_c + PI = q_o * w_o`.
-    ///
-    /// `{w_l, w_r, w_o, w_4} = {a, b, c, d}`
+    /// Forces `q_m * (a * b) + q_4 * d + q_c + q_o * c = 0`
     // XXX: Maybe make these tuples instead of individual field?
     pub fn big_mul_gate(
         &mut self,
@@ -149,6 +147,72 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
         self.q_o.push(q_o);
         self.q_c.push(q_c);
         self.q_4.push(q_4);
+        self.q_arith.push(E::Fr::one());
+
+        self.q_range.push(E::Fr::zero());
+        self.q_logic.push(E::Fr::zero());
+        self.q_fixed_group_add.push(E::Fr::zero());
+        self.q_variable_group_add.push(E::Fr::zero());
+
+        if let Some(pi) = pi {
+            assert!(
+                self.public_inputs_sparse_store.insert(self.n, pi).is_none(),"The invariant of already having a PI inserted for this position should never exist"
+            );
+        }
+
+        self.perm.add_variables_to_map(a, b, c, d, self.n);
+
+        self.n += 1;
+
+        c
+    }
+
+    /// This gates turns on all the selctor polynomials to give users,
+    /// in some situations, the ability to use the extra selector for
+    /// more variables into additions.
+    ///
+    /// This type of gate is usually used when we need to have
+    /// the largest amount of performance and the minimum circuit-size
+    /// possible. Since it allows the end-user to setup all of the selector
+    /// coefficients.
+    ///
+    /// Equation: `(a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI + q_o * c = 0`.
+    /// `d` will be set to zero if not provided.
+    ///
+    /// ### Returns
+    /// `c`
+    pub fn big_arith_gate(
+        &mut self,
+        a: Variable,
+        b: Variable,
+        c: Variable,
+        d: Option<Variable>,
+        q_m: E::Fr,
+        q_l: E::Fr,
+        q_r: E::Fr,
+        q_o: E::Fr,
+        q_c: E::Fr,
+        q_4: E::Fr,
+        pi: Option<E::Fr>,
+    ) -> Variable {
+        // Check if advice wire has a value
+        let d = match d {
+            Some(var) => var,
+            None => self.zero_var,
+        };
+
+        self.w_l.push(a);
+        self.w_r.push(b);
+        self.w_o.push(c);
+        self.w_4.push(d);
+
+        // Add selector vectors
+        self.q_m.push(q_m);
+        self.q_o.push(q_o);
+        self.q_c.push(q_c);
+        self.q_4.push(q_4);
+        self.q_l.push(q_l);
+        self.q_r.push(q_r);
         self.q_arith.push(E::Fr::one());
 
         self.q_range.push(E::Fr::zero());
@@ -302,6 +366,64 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
         let c = self.add_input(c_eval);
 
         self.big_mul_gate(a, b, c, Some(d), q_m, q_o, q_c, q_4, pi)
+    }
+
+    /// Adds a [`StandardComposer::big_arith_gate`] with the left, right
+    /// , fourth inputs and corresponding coefficients, computing & returning
+    /// the output (result) [`Variable`] and adding the corresponding arith
+    /// constraint.
+    ///
+    /// This type of gate is usually used when we don't need to have
+    /// the largest amount of performance and the minimum circuit-size
+    /// possible, since it defaults set `q_o` to `-1` to reduce the verbosity.
+    ///
+    /// Equation: `(a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI = c`
+    /// ### Returns
+    /// `c`
+    pub fn big_arith(
+        &mut self,
+        q_m: E::Fr,
+        a: Variable,
+        b: Variable,
+        q_l: E::Fr,
+        q_r: E::Fr,
+        q_4_d: Option<(E::Fr, Variable)>,
+        q_c: E::Fr,
+        pi: Option<E::Fr>,
+    ) -> Variable {
+        // check if advice wire is available
+        let (q_4, d) = match q_4_d {
+            Some((q_4, d)) => (q_4, d),
+            None => (E::Fr::zero(), self.zero_var),
+        };
+
+        // compute output wire
+        let a_eval = self.variables[&a];
+        let b_eval = self.variables[&b];
+        let d_eval = self.variables[&d];
+
+        let c_eval = (q_m * a_eval * b_eval)
+            + (q_l * a_eval)
+            + (q_r * b_eval)
+            + (q_4 * d_eval)
+            + q_c
+            + pi.unwrap_or_default();
+
+        let c = self.add_input(c_eval);
+
+        self.big_arith_gate(
+            a,
+            b,
+            c,
+            Some(d),
+            q_m,
+            q_l,
+            q_r,
+            -E::Fr::one(),
+            q_c,
+            q_4,
+            pi,
+        )
     }
 }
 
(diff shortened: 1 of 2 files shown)


tsunrise:
@LukePearson1 I ported the `big-arith-gate` from my old fork of dusk plonk. Can you take a look when you have time, thanks!

stechu:
To provide a bit context, this enables a constraint efficient Pedersen implementation:

spec: https://hackmd.io/jI0TLGF7QVWwxIqRNKqafw 
code: https://github.com/Manta-Network/Plonk-Prototype/pull/2

We will submit the Pedersen hash to the upstream once the code gets in a better shape.

CPerezz:
First of all, thanks for your contribution @tsunrise and @stechu .

We would appreciate (no need for this time) that you file an issue before you add any PR's to the repo. So that the work that you're planning to do can be discussed and we avoid scenarios where it ends up not being merged and resulting in a waste of time.

Aside from that, will review ASAP :smile:

tsunrise:
> First of all, thanks for your contribution @tsunrise and @stechu .
> 
> We would appreciate (no need for this time) that you file an issue before you add any PR's to the repo. So that the work that you're planning to do can be discussed and we avoid scenarios where it ends up not being merged and resulting in a waste of time.
> 
> Aside from that, will review ASAP 😄

Thanks! Yeah sorry, looks that I totally forgot this step but will definitely do that for next time = )
Feel free to ping me if there's need for some significant changes, and I will work on that soon

src/constraint_system/arithmetic.rs
```diff
@@ -169,6 +167,72 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
...
+        self.q_range.push(E::Fr::zero());
+        self.q_logic.push(E::Fr::zero());
+        self.q_fixed_group_add.push(E::Fr::zero());
+        self.q_variable_group_add.push(E::Fr::zero());
+
+        if let Some(pi) = pi {
+            assert!(
+                self.public_inputs_sparse_store.insert(self.n, pi).is_none(),"The invariant of already having a PI inserted for this position should never exist"
+            );
+        }
```
CPerezz, line 227: Let's create an error for that.

Something like: `PIPositionAlreadyInUse` or something less verbose maybe :sweat_smile: 

And the check could be done inside `self.public_inputs_sparse_store.insert(self.n, pi)` fn which should then get a wrapper which constructs the error. 

I think would be more clean.

src/constraint_system/arithmetic.rs
```diff
@@ -169,6 +167,72 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
...
+    /// the largest amount of performance and the minimum circuit-size
+    /// possible. Since it allows the end-user to setup all of the selector
+    /// coefficients.
+    ///
+    /// Equation: `(a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI + q_o * c = 0`.
+    /// `d` will be set to zero if not provided.
+    ///
+    /// ### Returns
+    /// `c`
+    pub fn big_arith_gate(
```
CPerezz, line 184: Seems that you're trying to do the same thing is done in https://github.com/rust-zkp/ark-plonk/blob/81e19eafed0cbef1761b56c5e176b8168d3b5fe5/src/constraint_system/composer.rs#L258 but adding `d` and `q_4`.

It's fine for me, but I would then delete the `poly_gate` and leave this one.

I would like to have consensus on this.
Any opinions @bhgomes @LukePearson1 @davidnevadoc ?

src/constraint_system/arithmetic.rs
```diff
@@ -303,6 +367,54 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
...
+        q_m: E::Fr,
+        a: Variable,
+        b: Variable,
+        q_l: E::Fr,
+        q_r: E::Fr,
+        q_4_d: Option<(E::Fr, Variable)>,
+        q_c: E::Fr,
+        pi: Option<E::Fr>,
+    ) -> Variable {
+        let q_o = -E::Fr::one();
```
CPerezz: This assigment is not needed. You can directly add it in the `big_arith_gate` call and if you want, highlight with a comment that corresponds to `q_o`.

Clippy will probably complain about it.

CPerezz (changes requested):
A couple of nits/comments.

For the rest, I really like it! Thanks a lot! :heart:

src/constraint_system/arithmetic.rs
CPerezz: Actually I think I'm wrong here as `poly_gate` does not compute `c` inside. So we can leave it.

src/constraint_system/ecc/mod.rs
```diff
@@ -14,6 +14,7 @@ use ark_ec::{
     twisted_edwards_extended::GroupAffine, PairingEngine, TEModelParameters,
 };
 use core::marker::PhantomData;
+use std::ops::Neg;
```
davidnevadoc: This can be removed and then the uses of `.neg()` be replaced by `-`.

src/constraint_system/ecc/mod.rs
```diff
@@ -167,28 +168,54 @@ impl<E: PairingEngine, P: TEModelParameters<BaseField = E::Fr>>
...
+    fn condition_point_neg(
+        &mut self,
+        bit: Variable, 
+        point_b: Point<E, P>
+    ) -> Point<E, P> {
+        let x = point_b.x;
+        let y = point_b.y;
+
+        // negation of point (x, y) is (-x, y)
+        let x_neg = self.add((E::Fr::one().neg(), x), (E::Fr::zero(), self.zero_var), E::Fr::zero(), None);
```
davidnevadoc: Replace `E::Fr::one().neg()` by `- E::Fr::one()`

src/constraint_system/ecc/mod.rs
```diff
@@ -203,5 +230,52 @@ mod tests {
...
+            let bit_0 = composer.zero_var();
+
+            let point = GroupAffine::new(
+                E::Fr::from(10u64),
+                E::Fr::from(20u64),
+            );
+            let point_var = Point::new(composer.add_input(point.x), composer.add_input(point.y));
+
+            let neg_point = composer.condition_point_neg(bit_1, point_var);
+            composer.assert_equal_public_point(neg_point, point.neg());
```
davidnevadoc: Replace `point.neg()` by ` -point`.

src/constraint_system/arithmetic.rs
davidnevadoc: This function doesn't compute `c` either right? 

The function `big_arith` computes `c` and `big_arith_gate` just enforces the constraint:
(a*b)*q_m + a*q_l + b*q_r + d*q_4 + q_c + PI + q_o * c = 0

I think it makes sense to rename `big_arith_gate` to `big_poly_gate` as it does the same thing with an extra wire.
Then `poly_gate` could call internally `big_poly_gate` with the 4th wire set to 0.

What do you think @CPerezz @LukePearson1 @bhgomes ?

davidnevadoc (changes requested):
I added a couple of comments suggesting some improvements, otherwise great work!

Thanks a lot for the contribution @stechu and @tsunrise :smile:

src/constraint_system/arithmetic.rs
bhgomes: This should not be an error, it should remain a panic. Since it is the composer internally tracking the gate counter, it's actually an implementation error if the gate PI gets messed up, not the user's fault, in which case it's not a recoverable error. You can make the code more succinct by just refactoring this into its own function, something like:

```rust
fn insert_public_input(&mut self, pi: Option<F>) {
    if let Some(pi) = pi {
        assert!(...)
    }
}
```

src/constraint_system/arithmetic.rs
bhgomes: Yeah we should consolidate these gates, in the same way that the `big_add_gate`/`big_add`/`add_gate`/`add` family is constructed. We should make sure that all three of the existing families are consistent (`poly`, `add`, and `mul`).

src/constraint_system/arithmetic.rs
bhgomes: Also, whenever this panics, it's an indication of a failure in _another function_ not the current one. It's actually a nontrivial issue if this panic ever gets called. There should be a comment in the panic that says to file a bug. If we had more type-level checks we could get rid of this failure path completely.

src/constraint_system/arithmetic.rs
LukePearson1: You would need `poly`, `poly_gate` and 'big_arith`, `arith` 

and they would have to follow the same structure as before

LukePearson1 (commented):
Left some comments, we should ensure whenever a new gate is made which asserts the outcome i.e. Variable(5) + Variable(8) = 13, is accompanied by the option to have the prover create it for us: Variable(5) + Variable(8) = ?.

bhgomes:
> We would appreciate (no need for this time) that you file an issue before you add any PR's to the repo. So that the work that you're planning to do can be discussed and we avoid scenarios where it ends up not being merged and resulting in a waste of time.

In general, I find it hard to describe/discover the right changes (to myself) without actually editing the code anyway. The first language I tend to use to express new ideas is the programming language itself, it's only later that I can describe what I did in natural language. So I think we shouldn't be too strict on Issue first, Pull-request second, especially since the PR system on GitHub still has all the same discussion features as Issues. If there are major changes, then some pre-existing discussion should definitely occur, but I don't think it's a "waste of time" to write code first, even if it is never used.

src/constraint_system/arithmetic.rs
davidnevadoc: I see the naming convention as follows:
 - The methods that have a `_gate` sufix (`add_gate`, `mul_gate`...) do not compute values, they receive as input the relevant wire values and selectors of the gate and check that the equation holds.
The methods that do not have the  `_gate` sufix (`add`, `mul`...) compute the output value `w_o` and always generate a satisfied constraint.
 - The methods with  `big_` prefix receive a 4th wire, those without the prefix don´t.

I think `big_arith` and `big_arith_gate` fit in the same family as `poly_gate` so we should choose a common name for all of them: `poly` or `arith`.

bhgomes: I agree. +1 for naming it `arith`.

tsunrise:
Thanks for the review 😃! I will try to respond to all of them by tomorrow

src/constraint_system/arithmetic.rs
CPerezz: Filed https://github.com/rust-zkp/ark-plonk/issues/46

CPerezz:
> > We would appreciate (no need for this time) that you file an issue before you add any PR's to the repo. So that the work that you're planning to do can be discussed and we avoid scenarios where it ends up not being merged and resulting in a waste of time.
> 
> In general, I find it hard to describe/discover the right changes (to myself) without actually editing the code anyway. The first language I tend to use to express new ideas is the programming language itself, it's only later that I can describe what I did in natural language. So I think we shouldn't be too strict on Issue first, Pull-request second, especially since the PR system on GitHub still has all the same discussion features as Issues. If there are major changes, then some pre-existing discussion should definitely occur, but I don't think it's a "waste of time" to write code first, even if it is never used.

Even if you don't see it as a waste of time, You need to think about everyone and not just your point of view in that specific case. And you need to understand that people might have this perception. Therefore what I try is to prevent this things by first just adding an issue even if it just tries to explain the big scope of what the person is trying to do. 
There he/she can get help., suggestions, guidance and most importantly, red flags that prevent the work to not be merged and frustration. 
As after being 4 years in the open source world, I've seen many things, and I try to avoid the ones based on mine and others experiences.

I understand your POV and I have quite some PR's in a lot of repos that haven't landed in months and will probably never land. But I would like that anyone that contributes to this repo, feels satisfied with the result and with it's work being included on it. :smile:

src/constraint_system/arithmetic.rs
CPerezz: hmmm ye. I was thinking when I said that that we should panic above. Since the error propagation will cause an error anyway. But then the API would become a nighmare with `?` everywhere and it's tricky. So yes, you're right, I prefer the panic too!

bhgomes: It's not just a convenience thing, it's also semantically wrong to propagate the error here and not panic.

src/constraint_system/ecc/mod.rs
tsunrise: fixed

src/constraint_system/ecc/mod.rs
tsunrise: fixed

src/constraint_system/ecc/mod.rs
tsunrise: fixed

src/constraint_system/arithmetic.rs
tsunrise: fixed

tsunrise:
Pushed some fixes, and have some questions unresolved: 
- Do we want rename `poly_gate` to `arith_gate` for this PR? 
- I feel like `big_arith_gate` and `big_poly_gate` should be in the same module. What do you think? 
- Do we want to return `c` in `big_arith_gate`? (It's also a little bit weird that `poly_gate` is returning something, but modifying the API `poly_gate` is a breaking change so I'm not sure. )
- see https://github.com/rust-zkp/ark-plonk/pull/43#discussion_r758489902_: do you want to do it in this PR? 

Thanks for the review!

tsunrise:
> Left some comments, we should ensure whenever a new gate is made which asserts the outcome i.e. Variable(5) + Variable(8) = 13, is accompanied by the option to have the prover create it for us: Variable(5) + Variable(8) = ?.

Do you mean have checks on both `big_arith_gate` and `big_arith`? For now, the tests only use `big_arith`, but `big_arith` uses `big_arith_gate` anyway

CPerezz:
>     * see [Add `big_arith_gate` and conditional point negate #43 (comment)](https://github.com/rust-zkp/ark-plonk/pull/43#discussion_r758489902)_: do you want to do it in this PR?
> 
> 
> Thanks for the review!

Hi @tsunrise !
No worries. I'm working on that on another PR. 

My suggestion is that we merge your changes and I'll do the refactor on top of them. So that at the end, instead of standarizing the fn names we have the builder pattern which already generalizes over all the options. 

WDYT??


Also, I would appreciate if you could rebase and clean up a bit the history. Then I'll run the workflow and we're ready to merge. No need to change any fn names as they'll be superset by #46 most probably :smile:

tsunrise:
Hi @CPerezz 
Sounds good! I will do it tomorrow

Also, what is the PR merging strategy for this repo? I can manually rebase it and make this PR to a single commit if we are not using squash merge

CPerezz:
@tsunrise it's up to you. But If I do the squash merge. I'll be the one that selects the message of the merge. 

If it's not an issue for you, then it's fine. If you want specific commit messages to appear in the squash commit, then I would suggest a rebase. 

Whatever you prefer. I will basically remove most of the commit messages and leave only 
37d0446  and d9316ae probably.

tsunrise:
got it thanks! I will make a rebase later

tsunrise:
done!

LukePearson1 (approved):
LGTM! 

Await 1 more reviewer approval and feel free to merge.

CPerezz (approved):
LGTM

davidnevadoc (commented):
LGTM :+1:

davidnevadoc (approved):

Issue #62, opened 2021-12-17
Implement changes to license and aesthetics


LukePearson1:
It has been decided that the license should be change for the direction of the organisation. This will remain MPL. 

There should also be some readme changes.

PR #63, opened 2021-12-17
Alter files and license header

LukePearson1:
This solves #62


CPerezz (changes requested):
All OK.

But i would prefer to keep the example in the README. Otherways, cargo can not check the correctnesd of the example while running the tests.

bhgomes:
> But i would prefer to keep the example in the README. Otherways, cargo can not check the correctnesd of the example while running the tests.

The test is in the source already.

bhgomes:
@LukePearson1 If you add a file called `.rustfmt.toml` and add the line

```
license_template_path = "FILE_TEMPLATE"
```

and then create a file called `FILE_TEMPLATE` and add the header you've added to every file in the repo to this file, then `rustfmt` will do a check that all the source code files have the correct copyright header.

davidnevadoc (approved):

bhgomes (approved):

PR #77, opened 2021-12-28
Todd/is eq gate

### LINKED ISSUE #76: Include custom gates for hashing purposes
When verifying paths in a Merkle tree it is useful to use a `conditional_select` together with a bit based on whether a hash refers to the left/right sibling at a given level.  To generate validity proofs of paths I propose a gate `is_eq_gate` that has two input wires and an output wire whose value is constrained to be 1 if the input wires have equal values and 0 otherwise.  It seems natural for this to be accompanied by `is_zero` and `is_nonzero` gates that have a single input wire and an output wire whose value is constrained to be 1 if the input is zero/nonzero and 0 otherwise.

GhostOfGauss:
Adds some gates that can be useful in combination with conditional selection.  They are 

*   `is-zero-gate` Has one input variable and one output variable.  The output variable is constrained to have the value 1 if the input is zero and to have the value 0 if the input is non-zero.
*   `is-eq-gate` has two input variables and an output variable which is constrained to have the value 1 if the two inputs have equal values and 0 otherwise.

The gates were added in the `boolean` submodule of the `constraint_system` module.  This seemed logical to me but should be checked.  

closes #76


src/constraint_system/boolean.rs
```diff
@@ -57,27 +57,32 @@ where
...
 
         let a_inv = self.add_input(a_inv_value);
         let result = self.add_input(result_value);
 
         // Enforce constraints
-        let a_times_result = self.arithmetic_gate(|gate| gate.witness(a, result, None).mul(E::Fr::one()));
+        let a_times_result = self.arithmetic_gate(|gate| {
+            gate.witness(a, result, None).mul(E::Fr::one())
+        });
         self.assert_equal(a_times_result, self.zero_var());
```
lopeetall: This `assert_equal` isn't necessary.  The product of `a` and `result` is constrained to be zero by the `mul` gate itself if you use `gate.witness(a, result, self.zero_var())`

src/constraint_system/boolean.rs
```diff
@@ -57,27 +57,32 @@ where
...
-        let sum = self.arithmetic_gate(|gate| gate.witness(a_times_ainv, result, None).add(E::Fr::one(), E::Fr::one()));
+        let a_times_ainv = self.arithmetic_gate(|gate| {
+            gate.witness(a, a_inv, None).mul(E::Fr::one())
+        });
+        let sum = self.arithmetic_gate(|gate| {
+            gate.witness(a_times_ainv, result, None)
+                .add(E::Fr::one(), E::Fr::one())
+        });
         let one = self.add_input(E::Fr::one());
         self.assert_equal(sum, one);
```
lopeetall: The `assert_equal` function ought to be used to compare variables, rather than constrain the output of a gate to a constant. Usually you can accomplish the same thing in the gate itself by using the constant wire. See here: https://github.com/ZK-Garage/plonk/blob/8080e7dd6d31b5fe4ff04e7c8f5bb69c21eeb09c/src/constraint_system/arithmetic.rs#L74

src/constraint_system/boolean.rs
```diff
@@ -57,27 +57,32 @@ where
...
         // Enforce constraints
-        let a_times_result = self.arithmetic_gate(|gate| gate.witness(a, result, None).mul(E::Fr::one()));
+        let a_times_result = self.arithmetic_gate(|gate| {
+            gate.witness(a, result, None).mul(E::Fr::one())
+        });
         self.assert_equal(a_times_result, self.zero_var());
 
-        let a_times_ainv = self.arithmetic_gate(|gate| gate.witness(a, a_inv, None).mul(E::Fr::one()));
-        let sum = self.arithmetic_gate(|gate| gate.witness(a_times_ainv, result, None).add(E::Fr::one(), E::Fr::one()));
+        let a_times_ainv = self.arithmetic_gate(|gate| {
```
lopeetall: The gates beginning here at line 73 can all be squashed into a single gate if you enable fan_in_3. 

```
self.arithmetic_gate(|gate| {
    gate.witness(a, a_inv, self.zero_var())
        .mul(E::Fr::one())
        .fan_in_3(E::Fr::one(), result)
        .constant(-E::Fr:one())
    });
```
This single gate constrains `1 * a * a_inv + 1 * result  - 1 = 0` and also avoids using another unnecessary `assert_equal` gate in line 81.

If you don't want to use `fan_in_3` you can do this in two gates by keeping the `mul` gate as you had it, and putting `self.zero_var()` as the output variable in the following `add` gate and still avoid the unnecessary `assert_equal` gate.

lopeetall (changes requested):

src/constraint_system/boolean.rs
GhostOfGauss: Thanks for the tip, I see how to use the `w_o` option correctly now.  Unfortunately doing so causes my tests to fail.  The error seems to occur here [https://github.com/arkworks-rs/algebra/blob/7657bcb7280ffb112056d59f2817c484ee4a6a6d/poly/src/polynomial/univariate/dense.rs#L35](https://github.com/arkworks-rs/algebra/blob/7657bcb7280ffb112056d59f2817c484ee4a6a6d/poly/src/polynomial/univariate/dense.rs#L35)

when getting the degree of a polynomial during proof generation.  Any idea why using `w_o : None` and `assert_equal` works but combining them as you suggested wouldn't? The exact change I made is to replace lines 68-70 with   
`let a_times_result = self.arithmetic_gate(|gate| { gate.witness(a, result, Some(zero)).mul(E::Fr::one())});`

where `zero` is `self.zero_var()`  
  
Could the problem be that this creates a variable `a_times_result` which is never used by the circuit?

lopeetall:
Ok, this is weird. I'm looking into it. All I know so far is that the
`q_arith` vector is all 1s, so it's polynomial representation is constant.
Then ark poly commit refuses to commit with KZG bc the poly is degree 0.

lopeetall:
Ok, mystery solved. This is because: `add_dummy_constraints` adds 2 arith gates to every circuit, and your circuit contains 6 more arith gates. This creates a circuit with 8 gates, all arith. This circuit is never padded out with zeros because it has a length of a power of 2 already. So the `q_arith` selector ends up as all 1s and its polynomial representation is degree 0 which triggers an error in KZG.

The original circuit had length 9. So it gets padded out to 16 with 0s and `q_arith` does not end up constant, so it works.

Indeed, if you make all the changes I suggested the total circuit length drops to 6. Then the test passes again!

I'll file an issue for the dummy constraint problem. It should be an easy fix.

davidnevadoc (changes requested):
Thanks for the PR!

Once the changes suggested by @lopeetall are introduced I think it will be ready to merge.

GhostOfGauss:
@lopeetall Thanks so much for getting to the bottom of that!  I wouldn't have thought of that.  It explains why I could get one of the two tests to pass but not both.    
  
I must be having trouble counting gates, because even when implementing your suggestions I still receive this error.  My current workaround is to add additional dummy constraints in the `is_eq` part of the gadget.  I suppose these can be removed once the 2^k bug is worked out.

lopeetall:
Oh boy. This goes deeper. Now it is `q_4` that is throwing the commit error, even though it is not the zero polynomial. I did some experimenting and believe the bug is in `ark-poly-commit`. 

The monomial basis form for `q_4` has a lower degree than expected for the domain and so it's leading coefficient is 0. This should be fine. However `ark-poly-commit` doesn't like this when it checks the degree and it throws an error. This is an oversight from `ark-poly-commit` and the way they check polynomial degree.

lopeetall:
The gates themselves look good to me. The 2^k problem only happens in your tests of the circuit, each of which add an extra constraint.

The `add_dummy_constraints()` hack to get around the `ark-poly-commit` bug is only really needed in the tests. If you add the dummy constraints to the gadget composer inside the failing tests they will pass again. Then at least we can keep the gates themselves free from any hacks that we will have to remember to remove later.

src/constraint_system/boolean.rs
```diff
@@ -50,6 +51,51 @@ where
...
+    /// two input variables have equal values and whose value is 0 otherwise.
+    pub fn is_eq_gate(&mut self, a: Variable, b: Variable) -> Variable {
+        let difference = self.arithmetic_gate(|gate| {
+            gate.witness(a, b, None).add(E::Fr::one(), -E::Fr::one())
+        });
+        // TODO: These dummy constraints are a hack to avoid a 2^k issue,
+        // specifically the q_arith poly being constant.
+        // It should be safe to remove this dummy constraint once that issue
+        // is resolved.
+        self.add_dummy_constraints();
```
CPerezz: This should be removed. 
Dummy constraints are only required for tests.
Anyway, the Composer by default adds some always (when you instanciate it) but I belive after @davidnevadoc 's PR, we don't have enough. 

This should be fixed, but not by adding the dummy constraints here.

src/constraint_system/boolean.rs
```diff
@@ -50,6 +51,51 @@ where
...
+                .fan_in_3(E::Fr::one(), b)
+                .constant(-E::Fr::one())
+        });
+
+        b
+    }
+
+    /// A gate which outputs a variable whose value is 1 if the
+    /// two input variables have equal values and whose value is 0 otherwise.
+    pub fn is_eq_gate(&mut self, a: Variable, b: Variable) -> Variable {
```
CPerezz: There's already a fn that does this in this repo.

Check: https://github.com/ZK-Garage/plonk/blob/9e7c26539816bf33873cbeb51672da98f6a50e85/plonk-core/src/constraint_system/composer.rs#L327

So this can be removed, or you can add this fn in the `composer.rs` module calling it something like: `is_eq_with_output` or something along these lines.

CPerezz (changes requested):

src/constraint_system/boolean.rs
GhostOfGauss: Thanks for your comments.  Right, the `assert_equal` fn already exists but does not give the user any output variable.  That is the reason for this PR, so it cannot be removed.  Are you suggesting that I locate it in the composer.rs module instead?

src/constraint_system/boolean.rs
GhostOfGauss: From @lopeetall 's comment above:

> Ok, mystery solved. This is because: `add_dummy_constraints` adds 2 arith gates to every circuit, and your circuit contains 6 more arith gates. This creates a circuit with 8 gates, all arith. This circuit is never padded out with zeros because it has a length of a power of 2 already. So the `q_arith` selector ends up as all 1s and its polynomial representation is degree 0 which triggers an error in KZG.
> 
> The original circuit had length 9. So it gets padded out to 16 with 0s and `q_arith` does not end up constant, so it works.
> 
> Indeed, if you make all the changes I suggested the total circuit length drops to 6. Then the test passes again!
> 
> I'll file an issue for the dummy constraint problem. It should be an easy fix.

So yes, this is a bad way to fix the problem.  Until the true problem is fixed this works, however.

GhostOfGauss: I should add here that the only problem with lopetall's comment is that the total circuit length drops to 6 within `is-zero-gate` but seems to then be 8 in `is-eq-gate`, hence the need to change the total number of gates there

src/constraint_system/boolean.rs
CPerezz: Exactly. And that you rename it on a way on which it implies the return of a `Veriable` in opposite to the other fn, :)

src/constraint_system/boolean.rs
CPerezz: It should be easy to solve. And I would encourage you to do so. 

In order to solve it and make this PR clean in that regard, the solution would be to add the call to `add_dummy_constraints` 8 times when you instanciate a new `StandardComposer`.

This is done here:
https://github.com/ZK-Garage/plonk/blob/9e7c26539816bf33873cbeb51672da98f6a50e85/plonk-core/src/constraint_system/composer.rs#L221

LukePearson1:
This is proving to have an interesting bug. Issue #78 explains the problem. @CPerezz

src/constraint_system/boolean.rs
GhostOfGauss: Resolved by #83 , the extra dummy constraints have been removed.

src/constraint_system/boolean.rs
GhostOfGauss: Okay, all done

LukePearson1 (approved):
With all of the changes spoken about previously, this is good to go since PR #83 was merged.

lopeetall (approved):
Looks great!

PR #87, opened 2022-01-11
Lookup error fixing

LukePearson1:


LukePearson1:
Will merge these changes into `plonkup_prover` so we can work from there

PR #89, opened 2022-01-17
Minor suggestions: using ToConstraintField, using HomomorphicCommitment everywhere, and removing unused append_commitments

ghost:
Some extra cleanup that I didn't do in the last PR.

1. Instead of using special structs, use arkworks `ToConstraintField` for public inputs. This does have to use (unfortunately) Boxes or other dynamic dispatch (to get trait objects) but maybe this is better.
2. There were some places I had used `PolynomialCommitment<F, DensePolynomial<F>>` instead of the `HomomorphicCommitment<F>` trait, which is slightly inconsistent. This makes everything consistent, using `PrimeField` as well.
3. Removes the unused `append_commitments` option, since using `append` for every `impl CanonicalSerialize` worked well.

### DIFF
--- a/plonk-core/src/circuit.rs
+++ b/plonk-core/src/circuit.rs
@@ -16,6 +16,51 @@ use ark_ec::models::TEModelParameters;
 use ark_ff::{Field, PrimeField, ToConstraintField};
 use ark_serialize::*;
 
+/// Public Input Builder
+#[derive(derivative::Derivative)]
+#[derivative(Default(bound = ""))]
+pub struct PublicInputBuilder<F>
+where
+    F: PrimeField,
+{
+    input: Vec<F>,
+}
+
+impl<F> PublicInputBuilder<F>
+where
+    F: PrimeField,
+{
+    /// Creates a new PublicInputBuilder
+    pub fn new() -> Self {
+        Self { input: vec![] }
+    }
+    /// Pushes a new input `item` into `self`, returning `None` if `item` could
+    /// not be converted into constraint field elements.
+    pub fn add_input<T>(self, item: &T) -> Option<Self>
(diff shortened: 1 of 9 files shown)


bhgomes:
@joebebel 

> Instead of using special structs, use arkworks ToConstraintField for public inputs. This does have to use (unfortunately) Boxes or other dynamic dispatch (to get trait objects) but maybe this is better.

I would suggest a builder pattern like the following instead of dynamic dispatch:

```rust
/// Public Input Builder
#[derive(derivative::Derivative)]
#[derivative(Default(bound = ""))]
pub struct InputBuilder<F> 
where
    F: PrimeField,
{
    input: Vec<F>,
}

impl<F> InputBuilder<F>
where
    F: PrimeField,
{
    /// Pushes a new input `item` into `self`, returning `None` if `item` could not be converted
    /// into constraint field elements.
    pub fn push<T>(self, item: &T) -> Option<Self>
    where
        T: ToConstraintField<F>,
    {
        self.extend([item])
    }
    
    /// Extends the input buffer by appending all the elements of `iter` converted into constraint
    /// field elements, returning `None` if any of the conversions failed.
    pub fn extend<'t, T, I>(mut self, iter: I) -> Option<Self>
    where
        T: 't + ToConstraintField<F>,
        I: IntoIterator<Item = &'t T>,
    {
        for item in iter {
            self.input.append(&mut item.to_field_elements()?);
        }
        Some(self)
    }
    
    /// Returns the vector of input elements.
    pub fn finish(self) -> Vec<F> {
        self.input
    }
}
```

bhgomes (approved):

LukePearson1 (approved):
LGTM

PR #90, opened 2022-01-17
Fix rendering

mathcrypto:


CPerezz (approved):
LGTM. 

Would like a review from @davidnevadoc and @LukePearson1 before merging

PR #42, opened 2021-11-27
Add lookups

LukePearson1:
Resolves #41

PR #94, opened 2022-01-26
Refactor Custom Gates

### LINKED ISSUE #112: Improve explanation of hierarchical selector polynomials in Plonk book
We can consider that `q_arith` is hierarchically at a higher level. 
`q_l, q_r, q_c, q_m, q_o, q_4` could be considered wire selectors in arithmetic gates, and auxiliary values in other custom gates.
On the other hand `q_arith, q_logic, ...` are gate selectors which determine which gate is activated at any point in the circuit.

The explanation should be expanded.

_Originally posted by @davidnevadoc in https://github.com/ZK-Garage/plonk/pull/107#discussion_r796522298_

davidnevadoc:
Refactor custom gates in order to abstract their contributions to proof, verifier and prover keys as well as the quotient and linearisation polynomial.

The objective is to detect unused custom gates at circuit compile time and eliminate all their contributions. This in turn:

- Reduces computation complexity of quotient and linearisation poly
- Reduces size of keys and proofs

### DIFF
--- a/plonk-core/src/proof_system/linearisation_poly.rs
+++ b/plonk-core/src/proof_system/linearisation_poly.rs
@@ -122,9 +122,13 @@ where
 
     /// Add evaluation of poly at point if the label is not already
     /// in the set of evaluations
-    pub fn add(label: String, poly: DensePolynomial<F>, point: F) {
-        // TODO
-        unimplemented!()
+    pub fn add(&mut self, label: &str, poly: DensePolynomial<F>, point: F) {
+        if let Some(_l) = &self.vals.iter().find(|entry| entry.0 == label) {
+            return;
+        } else {
+            let eval = poly.evaluate(&point);
+            let _ = &self.vals.push((label.to_string(), eval));
+        }
     }
 }
 
--- a/plonk-core/src/proof_system/linearisation_poly.rs
+++ b/plonk-core/src/proof_system/linearisation_poly.rs
@@ -8,11 +8,11 @@ use crate::{
     error::Error,
     label_eval,
     proof_system::{
-        ecc::{CurveAddition, FixedBaseScalarMul},
-        logic::Logic,
-        range::Range,
+        ecc::{CAVals, CurveAddition, FBSMVals, FixedBaseScalarMul},
+        logic::{Logic, LogicVals},
+        range::{Range, RangeVals},
         widget::GateConstraint,
-        ProverKey, WitnessValues,
+        CustomValues, ProverKey, WitnessValues,
     },
     util::EvaluationDomainExt,
 };
@@ -24,13 +24,6 @@ use ark_poly::{
 };
 use ark_serialize::{
     CanonicalDeserialize, CanonicalSerialize, Read, SerializationError, Write,
-};
-
-use super::{
-    ecc::{CAVals, FBSMVals},
-    logic::LogicVals,
-    range::RangeVals,
-    CustomValues,
 };
 
 /// Subset of the [`ProofEvaluations`]. Evaluations at `z` of the
--- a/plonk-core/src/proof_system/linearisation_poly.rs
+++ b/plonk-core/src/proof_system/linearisation_poly.rs
@@ -5,13 +5,15 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use crate::{
+    constraint_system::WireData,
     error::Error,
+    label_eval,
     proof_system::{
         ecc::{CurveAddition, FixedBaseScalarMul},
         logic::Logic,
         range::Range,
         widget::GateConstraint,
-        GateValues, ProverKey,
+        ProverKey, WitnessValues,
     },
     util::EvaluationDomainExt,
 };
@@ -40,11 +42,11 @@ where
     pub quot_eval: F,
 }
 
-/// Subset of all of the evaluations. These evaluations
-/// are added to the [`Proof`](super::Proof).
+/// Subset of the [`ProofEvaluations`]. Evaluations at `z` of the
+/// wire polynomials
 #[derive(CanonicalDeserialize, CanonicalSerialize, derivative::Derivative)]
 #[derivative(Clone, Debug, Default, Eq, PartialEq)]
-pub struct ProofEvaluations<F>
+pub struct WireEvaluations<F>
 where
     F: Field,
 {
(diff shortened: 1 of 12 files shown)


davidnevadoc:
The original purpose of the PR was to detect unused custom gates and eliminate their contributions to proof and keys when a circuit is compiled. However it may be a good idea to review and merge this PR now in order to have as few conflicts as possible with the lookups PR (#42 ) when it is merged.
I also think that at this point it would be good to receive inputs about the changes made and open a discussion about what could be the best design to achieve this goal.

Some assumptions I have made:
 - I have not considered arithmetic gates as custom gates and think the should be non-removable. This is only a downside for circuits that have no arithmetic gates which I think are rare.  Moreover, when we switch the blinding method (issue #113 ) we will introduce arithmetic gates anyway.
 - Every custom gate has a selector polynomial. This is true at the moment but there may be better ways of going about gate selection, as @bhgomes pointed out (in https://github.com/ZK-Garage/plonk/issues/112#issuecomment-1029181120_)
- For the computation of the linearisation polynomial the selector polynomials are not chosen to be evaluated  and added to the proof (with the exception of `q_arith`). The rest of the polynomials needed are evaluated and added instead. This is not the best strategy (for optimizing number of evaluations in the proof) but it is simple and generic.

## Summary of changes

Regarding the widgets:
- The struct `GateValues` has been divided in `WitnessValues` and `CustomValues`
    - `WinessValues` is a struct that holds the values of the wires at a given point
    -  `CustomValues` is a trait that is implemented by each custom gate specific struct: 
      (`LogicVals`, `RangeVals`, `CAVals` and `FBSMVals`)
- Move away from the terms `right, left, out, fourth` for the wires in favor of `a,b,c,d` (for consistency with the rest of the library).

Regarding the linearisation polynomial
- The current version of the Plonk paper has an improved version of the lin poly which allows the removal of the lin poly evaluation in the proof. The code has been updated with this new version.
-  `Evaluations` struct has been removed. The improved version removes the need to pass around the quotient and lin polys evaluations so the struct was now obsolete (It was just `quot_eval` + `ProofEvaluations`). 
-  `ProofEvaluations` has been divided in 3 structs:
   -  `WireEvaluations`
   -  `PermutationEvaluations`
   -  `CustomEvaluations`

davidnevadoc:
These changes also solve the bug of the missing `q_arith` evaluation in the proof. Now it is considered part of the custom evaluations and it is properly added.

plonk-core/src/proof_system/linearisation_poly.rs
```diff
@@ -25,26 +26,18 @@ use ark_serialize::{
...
-where
-    F: Field,
-{
-    /// Proof-relevant Evaluations
-    pub proof: ProofEvaluations<F>,
-
-    /// Evaluation of the linearisation sigma polynomial at `z`.
-    pub quot_eval: F,
-}
+use super::{
```
markulf: The split into `use crate` an `use super` is a bit strange? `CustomValues` should be just after `WitnessValues`?

plonk-core/src/proof_system/widget/arithmetic.rs
```diff
@@ -43,23 +45,20 @@ where
 
 impl<F> ProverKey<F>
 where
-    F: FftField,
```
LukePearson1, line 46: With all the changes to different Field traits, as well as back and forth. Can we clarify the use of PrimeField over FFTField?

plonk-core/src/proof_system/linearisation_poly.rs
mathcrypto: > The split into `use crate` an `use super` is a bit strange? `CustomValues` should be just after `WitnessValues`?

I've noticed the same in the logic gate as well. Why is this @davidnevadoc ?

plonk-core/src/proof_system/widget/logic.rs
```diff
@@ -6,10 +6,23 @@
...
+use super::CustomValues;
+
+pub struct LogicVals<F>
+where
+    F: PrimeField,
+{
+    pub left_next: F,
+    pub right_next: F,
+    pub fourth_next: F,
+    pub constant_selector: F,
```
mathcrypto: where is ```left_selector``` and ```right_selector``` ?

davidnevadoc: They are not used in the logic gate so we don't add them here. 
The idea is that for each custom gate the struct that implement `CustomValues` hold only the values they need to compute their constraints.

plonk-core/src/proof_system/linearisation_poly.rs
davidnevadoc: You're right, it's a bit odd. I added these automatically with the IDE and didn't notice the inconsistency. 
I'll fix it right away! :)

plonk-core/src/proof_system/widget/logic.rs
mathcrypto: Ok that explains it, thank you

CPerezz (approved):
We already reviewed part of this in our calls!

Nice job mate! LGTM

plonk-core/src/proof_system/widget/logic.rs
```diff
@@ -6,10 +6,23 @@
 
 //! Logic Gates
 
-use crate::proof_system::widget::{GateConstraint, GateValues};
+use crate::proof_system::widget::{GateConstraint, WitnessValues};
```
mathcrypto: This is where I was referring @davidnevadoc about the usage of ```crate``` and ```super```

plonk-core/src/proof_system/widget/range.rs
```diff
@@ -6,10 +6,21 @@
 
 //! Range Gate
 
-use crate::proof_system::{GateConstraint, GateValues};
+use crate::proof_system::{GateConstraint, WitnessValues};
```
mathcrypto: Same thing here @davidnevadoc

LukePearson1 (approved):
Bar the comment from me, if @CPerezz and you had this conversation then it looks good to merge 👍

plonk-core/src/proof_system/widget/arithmetic.rs
bhgomes: `PrimeField` implements `FftField` which implements `Field`. The most general API is `Field`, `FftField`, or `PrimeField` in increasing specificity. I recommend `Field` since it is the most general, and we only need `FftField` and `PrimeField` in some places.

PR #110, opened 2022-02-01
Benchmarks

iquerejeta:
Make benchmarks generic, and provide two instantiations. BLS12_381 with KZG10 and IPA.


benches/plonk.rs
```diff
@@ -65,34 +66,53 @@ where
...
+    F: PrimeField,
+    P: TEModelParameters<BaseField = F>,
+    HC: HomomorphicCommitment<F>,
+{
     let label = b"ark".as_slice();
 
-    const MINIMUM_DEGREE: usize = 5;
-    const MAXIMUM_DEGREE: usize = 19;
+    const MINIMUM_DEGREE: usize = 2;
+    const MAXIMUM_DEGREE: usize = 5;
```
davidnevadoc: Let's use the original values to `5` to `19`

davidnevadoc (changes requested):
Great job! 
Once the min and max degrees are changed back to the originals this is ready to be merged. :+1:

benches/plonk.rs
iquerejeta: woops

PR #111, opened 2022-02-01
Fix CI

davidnevadoc:
Solve the current CI check fails

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: macos-latest 
+    runs-on: macos-latest
     steps:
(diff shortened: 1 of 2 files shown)


davidnevadoc:
The error in `intra-doc-links` is due to  [this issue](https://github.com/rust-lang/rust/issues/93476). Probably will be fixed in the next copule days.

davidnevadoc:
We should consider fixing a version `rust-toolchain` and update it manually in order to avoid these issues.

CPerezz (approved):
LGTM!!!!!!

LukePearson1 (approved):
Thanks for finding the issue. @davidnevadoc

mathcrypto (approved):

Issue #109, opened 2022-01-31
Hashing trait


iquerejeta:
With the progress of the Poseidon implementation #104 , it would be nice to begin a discussion on how we could implement the hash trait(s). Hopefully, this will also steer the discussion towards the gadgets design. I will eventually draft a PR.

My idea is to break the hashing family functionality into different traits: 
* `CollisionResistantHashing`
* `PseudoRandomFunction`
* `PreImageResistantHashing` (and 2nd-pre image)

This would allow us to use hash functions which might only require collision resistance (certain use cases of merkle trees), cryptographically safe hash functions (collision + preimage + 2nd preimage resistance), or pseudo random functionality. 

Was discussing with @CPerezz that it might make sense to also separate prime field hash functions and those that operate over binary fields. I agree with that, and in what concerns the prime field hash functions I would also make explicit trait distinctions for the capacity of the functions, as these might require different parametrisation. So rather than going with [this type of trait](https://github.com/webb-tools/arkworks-gadgets/blob/master/arkworks-plonk-circuits/src/poseidon/poseidon.rs#L30) definition (with `hash` and `hash_two` part of the same trait), I would rather follow [this style](https://github.com/arkworks-rs/crypto-primitives/blob/main/src/crh/constraints.rs#L27) (making an explicit distinction).  

Any thoughts and comments welcome 👍

markulf:
Your preferred style might seem more verbose as the traits are otherwise quite similar, but these seem to be different use-cases and it might make sense to treat them seperately.

Two observations, the arkworks trait, i.e. the second design, seems to hide the `composer: &mut StandardComposer<F, P>` but requires repeated input of parameters, which are taken from `self` in the first design.

In general I like the generic nature of arkworks and unless it comes at huge costs, either in performance or complicated code, it might be good to stay close to their design. This is likely a much wider discussion than this issue.

A side note, in fact this is the trait for the [gadget](https://github.com/arkworks-rs/crypto-primitives/blob/main/src/crh/constraints.rs), while what you pointed to might just be for native hash functions.

iquerejeta:
Thanks for the comments. You are right, I missed the correct link. Edited with the one of the gadget now 👍  

Regarding the composer, my first thought would be that the hash gadgets take a mutable reference to the standard composer. But I guess this question depends on how we decide to design gadgets in general, and not only hashing. 

[...]

PR #114, opened 2022-02-04
Poseidon hash

### LINKED ISSUE #104: Poiseidon Hash Implementation
Opening this to alert everyone that I am in the process of adapting Manta's poseidon implementation here (https://github.com/Manta-Network/Plonk-Prototype/pull/3)  to be ready to enter the plonk-hashing module.  

This implementation includes [filecoin's optimizations](https://github.com/filecoin-project/neptune) (see [here ](https://hackmd.io/8MdoHwoKTPmQfZyIKEYWXQ?view#Optimized-Poseidon-Hash)too) as well as further optimizations that @tsunrise added.  

I will link this issue to a PR over the coming days.

GhostOfGauss:
This PR adds our first hash function--the optimized Poseidon hash, complete with parameter generation written in Rust. Credit goes to @tsunrise and @BoyuanFeng for writing this.  The contents of this PR is adapted from their implementation [here](https://github.com/Manta-Network/Plonk-Prototype/pull/3).  (See also [this spec](https://hackmd.io/8MdoHwoKTPmQfZyIKEYWXQ?view#Optimized-Poseidon-Hash)) The biggest change is the addition of a `HashFunction` trait and making hashing stateless.

Some work remains, but I'm opening this draft to start collecting feedback.  What remains is documentation and testing.

Feedback is needed on the overall interface with hash functions.  The `HashFunction` trait is a starting point for discussion. This topic was opened in Issue #109 and discussion should continue there.  

This will close #104


LukePearson1:
Is this PR ready for review @GhostOfGauss? If so, please convert from draft.

GhostOfGauss:
No, not ready yet

plonk-hashing/src/lib.rs
```diff
@@ -6,7 +6,12 @@
...
 
-#![cfg_attr(not(any(feature = "std", test)), no_std)]
+// #![cfg_attr(not(any(feature = "std", test)), no_std)]
 #![cfg_attr(doc_cfg, feature(doc_cfg))]
 #![forbid(rustdoc::broken_intra_doc_links)]
-#![forbid(missing_docs)]
+// #![forbid(missing_docs)]
+
+#[macro_use]
+pub extern crate ark_std;
```
bhgomes: We should not depend on `ark_std` and use the `core` and `alloc` libraries instead.

plonk-hashing/src/poseidon/poseidon.rs
```diff
@@ -0,0 +1,584 @@
...
+            *val = c.arithmetic_gate(|g| {
+                g.witness(state[0], state[j], None)
+                    .add(matrix.v_rest[j - 1], F::one())
+            });
+        }
+        *state = result;
+    }
+}
+
+trait HashFunction<const WIDTH: usize, COM = ()> {
```
bhgomes, line 557: This should be arity, not width since the notion of "width" is a poseidon-only thing.

plonk-hashing/src/poseidon/poseidon.rs
```diff
@@ -0,0 +1,584 @@
...
+    type Input;
+    type Output;
+
+    // The input ought to have size ARITY, but const generic expressions aren't
+    // allowed yet
+    fn hash(&self, input: &[Self::Input], compiler: &mut COM) -> Self::Output;
```
bhgomes, line 563: We can use const-generics. See this example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cc008c861ad5c5455d1ca26622284a4b.

GhostOfGauss: But can we use const-generic _expressions_?  The issue I was referring to is that as you said above, `HashFunction` trait should have the generic `ARITY` but within `PoseidonSpec` we need to arrays whose size is `ARITY + 1`.  That produces "error: generic parameters may not be used in const operations" with the suggestion to "use `#![feature(generic_const_exprs)]` to allow generic const expressions"

plonk-hashing/src/poseidon/poseidon.rs
GhostOfGauss: Agreed, but see comment below regarding const generic expressions

plonk-hashing/src/lib.rs
GhostOfGauss: Okay, took care of this.  There were two spots ([here](https://github.com/ZK-Garage/plonk/blob/097353b1eec48eb3bb083c44342b8535a1e3d335/plonk-hashing/src/poseidon/mds.rs#L3) and [here ](https://github.com/ZK-Garage/plonk/blob/097353b1eec48eb3bb083c44342b8535a1e3d335/plonk-hashing/src/poseidon/preprocessing.rs#L4)) where Manta's implementation uses `ark_ff::vec` and I was unsure whether to remove these.  It compiles fine without these imports--can you confirm that it's safe to remove them?

plonk-hashing/src/poseidon/poseidon.rs
bhgomes: The notion of width is internal to Poseidon, so the way it's represented in memory is an implementation detail, not something that is exposed to the interface. Using arrays would be nice but since the feature is unstable, the other options are to use a `Vec` for the state vector (of length `ARITY + 1`) or to separate the state into two parts like this:

```rust
type State = (F, [F; ARITY]);
```

In either case, the input to the hash function is of the form `[F; ARITY]`, the rest of the state is not part of the interface.

NB: In my implementation in `manta-rs` I use `Vec` for simplicity but for the production circuit I'm going to try and use `State` as above.

plonk-hashing/src/lib.rs
bhgomes: The `ark_ff::vec` should be removed in favor of `alloc::vec`. It's just a re-export.

plonk-hashing/src/poseidon/poseidon_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+    //     let digest4 = p4.hash_in_mode(OptimizedDynamic);
+
+    #[test]
+    fn compare_with_neptune() {
+        const ARITY: usize = 2;
+        const WIDTH: usize = ARITY + 1;
+        type NepArity = generic_array::typenum::U2;
+
+        let (nep_consts, ark_consts) =
+            collect_neptune_constants::<NepArity>(Strength::Standard);
```
markulf, line 496: `collect_neptune_constants` is missing, as well as neptune dev_dependencies? Nice to have these tests.

plonk-hashing/src/poseidon/poseidon_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+        let inputs_var =
+            inputs.iter().map(|x| c.add_input(*x)).collect::<Vec<_>>();
+        let mut poseidon_circuit =
+            PoseidonRef::<_, PlonkSpecRef, WIDTH>::new(&mut c, param);
+        inputs_var.iter().for_each(|x| {
+            let _ = poseidon_circuit.input(*x).unwrap();
+        });
+        let plonk_hash = poseidon_circuit.output_hash(&mut c);
+
+        c.check_circuit_satisfied();
```
markulf, line 386: `check_circuit_satisfied` requires `#[cfg(feature = "trace")]`

plonk-hashing/src/poseidon/poseidon_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+
+impl<COM, S: PoseidonRefSpec<COM, WIDTH>, const WIDTH: usize>
+    PoseidonRef<COM, S, WIDTH>
+{
+    pub fn new(
+        c: &mut COM,
+        constants: PoseidonConstants<S::ParameterField>,
+    ) -> Self {
+        let mut elements = S::zeros(c);
+        elements[0] = S::alloc(c, constants.domain_tag);
```
markulf, line 157: This loads `constants.domain_tag` into a witness variable using `c.add_input`, but how do we constrain the witness to correspond to the actual constants?

plonk-hashing/src/poseidon/constraints.rs
```diff
@@ -0,0 +1,473 @@
+//! Library independent specification for field trait.
```
tsunrise, line 1: I don't think this file is needed here

plonk-hashing/src/poseidon/poseidon.rs
tsunrise: Just a side note, this poseidon implementation use `WIDTH` (`ARITY + 1`) instead of `ARITY`. We can change it to `ARITY` once `generic_const_exprs` is stable. 
@bhgomes I guess we have to use slice instead of const generics? because
- as @GhostOfGauss said, cannot evaluate `WIDTH - 1` at compile time
- compiler will probably optimize the overhead out?

plonk-hashing/src/poseidon/poseidon_ref.rs
tsunrise: Good point... a naive way is that for the input phase we can just use an `zero_var` for `element[0]` and keep track of domain tag, and then we can handle this in the first round of `add_round_constants` and s-box.

plonk-hashing/src/poseidon/poseidon_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+
+        c.check_circuit_satisfied();
+        println!(
+            "circuit size for WIDTH {} poseidon: {}",
+            WIDTH,
+            c.circuit_size()
+        )
+    }
+
+    #[test]
```
tsunrise, line 399: we do not need to test on `r1cs`. 

(Indeed, we should remove everything related `r1cs` here. It's not relevant. )

plonk-hashing/src/poseidon/poseidon_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+            let _ = poseidon.input(Fr::rand(&mut rng)).unwrap();
+        });
+        let _ = poseidon.output_hash(&mut ());
+    }
+
+    use crate::tests::{
+        conversion::cast_field,
+        neptune_hyper_parameter::collect_neptune_constants,
+    };
+    use neptune::{
```
tsunrise, line 464: We should use some hardcoded test instead of relying on `neptune` library.

plonk-hashing/src/poseidon/poseidon_ref.rs
tsunrise: This code is from `plonk_prototype` repo and we temporarily used neptune library as a reference to check the correctness of our code. 

Anyway, we should not even import `neptune` library here. We need to hardcoded tests.

plonk-hashing/src/poseidon/round_constant.rs
```diff
@@ -0,0 +1,171 @@
...
+    append_bits(&mut init_sequence, 10, r_p); // Bits 40-49
+    append_bits(&mut init_sequence, 30, 0b111111111111111111111111111111u128); // Bits 50-79
+
+    let mut grain = GrainLFSR::new(init_sequence, field_size);
+    let mut round_constants: Vec<F> = Vec::new();
+
+    match field {
+        1 => {
+            for _ in 0..num_constants {
+                while {
```
tsunrise, line 43: use `F::from_le_bytes_mod_order` instead of `while` loop here. See https://github.com/arkworks-rs/sponge/blob/51d6fc9aac1fa69f44a04839202b5de828584ed8/src/poseidon/grain_lfsr.rs#L135

tsunrise (commented):
Thanks

plonk-hashing/src/poseidon/poseidon.rs
bhgomes: @tsunrise To make it equivalent to the array case in terms of efficiency, we only need to keep all the data on the stack, so we can use something like `(F, [F; ARITY])` as the state vector. This also preserves the input memory format since the state enters the algorithm in these two parts anyway, so copy elision will remove some overhead here that you would have if you had to copy everything into a `[F; ARITY + 1]` which is more nontrivial to optimize. Maybe the Rust optimizer is very good but I wouldn't rely on optimizations that are not explicitly based on the language semantics like ownership and moving.

plonk-hashing/src/poseidon/poseidon_ref.rs
markulf: You mean hardcoded test vectors. Sounds good.

plonk-hashing/src/poseidon/constants.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub domain_tag: F,
+    pub full_rounds: usize,
+    pub half_full_rounds: usize,
+    pub partial_rounds: usize,
+}
+
+impl<F: PrimeField> PoseidonConstants<F> {
+    /// Generate all constants needed for poseidon hash of specified
+    /// width.  Note that WIDTH = ARITY + 1
+    pub fn generate<const WIDTH: usize>() -> Self {
```
iquerejeta, line 26: Would it make sense to declare WIDTH as a constant to the structure rather than the function? This way we ensure that the poseidon trait and the constants used are initialised for the same width.

tsunrise:
@GhostOfGauss The round constant generation code in Manta's plonk prototype repo was incorrect, and I fixed that so it now passes the hardcoded tests. When you have some time can you sync those changes here?

plonk-hashing/src/poseidon/poseidon_ref.rs
tsunrise: Those hardcoded test vectors are now here: https://github.com/Manta-Network/Plonk-Prototype/blob/d2e25aae7ad4526791f4543cebf1e4df5fbf14ac/src/poseidon/poseidon_ref.rs#L380-L415. Those test vectors come from my modified variants of the original sage script. @GhostOfGauss You can add this test here  remove any dependency on Neptune

plonk-hashing/src/poseidon/poseidon.rs
```diff
@@ -0,0 +1,582 @@
...
+        let post_round_keys = constants
+            .compressed_round_constants
+            .iter()
+            .skip(*const_offset)
+            .take(to_take);
+
+        if !last_round {
+            let needed = *const_offset + to_take;
+            assert!(
+                needed <= constants.compressed_round_constants.len(),
```
iquerejeta, line 106: This overflow could happen in any round, not only the last, right? Might be missing some check performed on the number of constants though. But maybe bounding the constants with the Width (as suggested in a different comment) might help in avoiding these type of overflows.

plonk-hashing/src/poseidon/mds.rs
```diff
@@ -0,0 +1,257 @@
...
+        let v = m[0][1..].to_vec();
+        let w = m.iter_rows().skip(1).map(|column| column[0]).collect();
+        (v, w)
+    }
+}
+
+/// A `SparseMatrix` is specifically one of the form of M''.
+/// This means its first row and column are each dense, and the interior matrix
+/// (minor to the element in both the row and column) is the identity.
+/// We will pluralize this compact structure `sparse_matrixes` to distinguish
```
iquerejeta, line 121: Naming could be confusing. What about `sparse_factorized` instead of `sparse_matrixes`?

plonk-hashing/src/poseidon/poseidon.rs
```diff
@@ -0,0 +1,582 @@
...
+
+        for (j, val) in result.iter_mut().enumerate().skip(1) {
+            // for each j, result[j] = state[j] + state[0] * v_rest[j-1]
+
+            // Except for first row/column, diagonals are one.
+            *val = Self::add(c, val, &state[j]);
+            //
+            // // First row is dense.
+            let tmp = Self::muli(c, &state[0], &matrix.v_rest[j - 1]);
+            *val = Self::add(c, val, &tmp);
```
iquerejeta, line 270: might be missing important points of how constraints are built, but would it be possible to merge both additions into one? First one is after all an addition of zero
```
*val = Self::add(c, state[j], &tmp);
```

plonk-hashing/src/poseidon/mds.rs
```diff
@@ -0,0 +1,257 @@
...
+/// from `sparse_matrices` from which they are created.
+#[derive(Debug, Clone, PartialEq)]
+pub struct SparseMatrix<F: PrimeField> {
+    /// `w_hat` is the first column of the M'' matrix. It will be directly
+    /// multiplied (scalar product) with a row of state elements.
+    pub w_hat: Vec<F>,
+    /// `v_rest` contains all but the first (already included in `w_hat`).
```
iquerejeta, line 128:
```suggestion
    /// `v_rest` contains the first row except for the first value (already included in `w_hat`).
```

plonk-hashing/src/poseidon/mds.rs
```diff
@@ -0,0 +1,257 @@
...
+use ark_ff::PrimeField;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MdsMatrices<F: PrimeField> {
+    pub m: Matrix<F>,
+    pub m_inv: Matrix<F>,
+    pub m_hat: Matrix<F>,
+    pub m_hat_inv: Matrix<F>,
+    pub m_prime: Matrix<F>,
+    pub m_double_prime: Matrix<F>,
```
iquerejeta, line 13: can't this one be defined as sparse?

iquerejeta (commented):
Some comments from a first pass.

plonk-hashing/src/poseidon/poseidon.rs
```diff
@@ -0,0 +1,582 @@
...
+
+// TODO: reduce duplicate code with `poseidon_ref`
+pub trait PoseidonSpec<COM, const WIDTH: usize> {
+    type Field: Debug + Clone;
+    type ParameterField: PrimeField;
+
+    fn output_hash(
+        c: &mut COM,
+        elements: &[Self::Field; WIDTH],
+        constants: &PoseidonConstants<Self::ParameterField>,
```
drewstone, line 22: Could we consider adding an index for the output element here? Or maybe some understanding on why the 2nd element is selected and not the 1st (0-indexed element).

plonk-hashing/src/poseidon/poseidon.rs
tsunrise: Yes, you are right. I've updated the code in the prototype repo [here](https://github.com/Manta-Network/Plonk-Prototype/blob/30ee9778f4d905b1f48e3ba4086bcc96d32dd061/src/poseidon/poseidon.rs#L500-L516). I believe those changes will soon be synced to this PR.

PR #116, opened 2022-02-09
Fix katex syntax

icemelon:
Previously one set of equations wasn't rendered correctly due to some syntax error.


markulf (approved):
This is great! Thanks a lot!

LukePearson1 (approved):
This looks perfect to me. Thanks a lot @icemelon!

PR #117, opened 2022-02-11
Update blinder method

### LINKED ISSUE #113: Update blinder method
We are currently blinding the wire and permutation polynomials with the method proposed in the Plonk paper. This method consists of adding a multiple (of the appropriate degree) of the vanishing polynomial.
Having 4 wires (instead of 3) makes this method quite inconvenient  as it raises the degree of the quotient polynomial above `4n`. This forces us to use an `8n` sized domain for the FFTs which negatively impacts performance.

The current blinders will be switched with the method described in [this article](https://mirprotocol.org/blog/Adding-zero-knowledge-to-Plonk-Halo) as suggested by @lopeetall .

markulf:
Introduce blinders that do not increase the size of the evaluation domain.

### DIFF
--- a/plonk-core/src/proof_system/quotient_poly.rs
+++ b/plonk-core/src/proof_system/quotient_poly.rs
@@ -52,54 +52,54 @@ where
     F: PrimeField,
     P: TEModelParameters<BaseField = F>,
 {
-    let domain_8n = GeneralEvaluationDomain::<F>::new(8 * domain.size())
+    let domain_4n = GeneralEvaluationDomain::<F>::new(4 * domain.size())
         .ok_or(Error::InvalidEvalDomainSize {
-        log_size_of_group: (8 * domain.size()).trailing_zeros(),
+        log_size_of_group: (4 * domain.size()).trailing_zeros(),
         adicity:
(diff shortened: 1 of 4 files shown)


plonk-core/src/proof_system/quotient_poly.rs
```diff
@@ -235,38 +235,38 @@ where
...
+    z_eval_4n: &[F],
     alpha: F,
     beta: F,
     gamma: F,
 ) -> Result<Vec<F>, Error>
 where
     F: PrimeField,
 {
-    let domain_8n = GeneralEvaluationDomain::<F>::new(8 * domain.size())
+    let domain_4n = GeneralEvaluationDomain::<F>::new(8 * domain.size())
```
davidnevadoc, line 250: The size is now `4 * domain.size()`

plonk-core/src/proof_system/quotient_poly.rs
```diff
@@ -235,38 +235,38 @@ where
...
         .ok_or(Error::InvalidEvalDomainSize {
         log_size_of_group: (8 * domain.size()).trailing_zeros(),
```
davidnevadoc, line 252: same here

davidnevadoc:
This PR has been moved to #118

PR #118, opened 2022-02-17
New blinders

### LINKED ISSUE #113: Update blinder method

davidnevadoc:
Closes #113

### DIFF
--- a/plonk-core/src/constraint_system/composer.rs
+++ b/plonk-core/src/constraint_system/composer.rs
@@ -787,7 +787,7 @@ mod test {
         F: PrimeField,
         P: TEModelParameters<BaseField = F>,
     {
-        // NOTE: Circuit size is n+3 because
+        // NOTE: Circuit size is n+4 because
         // - We have an extra gate which forces the first witness to be zero.
         //   This is used when the advice wire is not being used.
         // - We have two gates which add random values to blind the wires.
--- a/plonk-core/src/constraint_system/composer.rs
+++ b/plonk-core/src/constraint_system/composer.rs
@@ -21,6 +21,7 @@ use ark_ec::{models::TEModelParameters, ModelParameters};
 use ark_ff::PrimeField;
 use core::marker::PhantomData;
 use hashbrown::HashMap;
+use rand::{CryptoRng, RngCore};
 
 /// The StandardComposer is the circuit-builder tool that the `plonk` repository
 /// provides to create, stored and transformed circuit descriptions
@@ -212,7 +213,7 @@ where
             composer.add_witness_to_circuit_description(F::zero());
 
         // Add dummy constraints
-        composer.add_blinding_factors();
+        composer.add_blinding_factors(&mut rand::rngs::OsRng);
 
         composer
     }
@@ -530,15 +531,18 @@ where
     /// and permutation polynomials.
     /// All gate selectors are turned off to guarantee the constraints
     /// are still satisfied.
-    pub fn add_blinding_factors(&mut self) {
+    pub fn add_blinding_factors<R>(&mut self, rng: &mut R)
+    where
+        R: CryptoRng + RngCore + ?Sized,
+    {
         let mut rand_var_1 = Variable::default();
         let mut rand_var_2 = Variable::default();
         // Blinding wires
         for _ in 0..2 {
-            rand_var_1 = self.add_input(F::rand(&mut rand::rngs::OsRng));
-            rand_var_2 = self.add_input(F::rand(&mut rand::rngs::OsRng));
-            let rand_var_3 = self.add_input(F::rand(&mut rand::rngs::OsRng));
-            let rand_var_4 = self.add_input(F::rand(&mut rand::rngs::OsRng));
+            rand_var_1 = self.add_input(F::rand(rng));
+            rand_var_2 = self.add_input(F::rand(rng));
+            let rand_var_3 = self.add_input(F::rand(rng));
+            let rand_var_4 = self.add_input(F::rand(rng));
 
             self.w_l.push(rand_var_1);
             self.w_r.push(rand_var_2);
--- a/plonk-core/src/constraint_system/composer.rs
+++ b/plonk-core/src/constraint_system/composer.rs
@@ -535,8 +535,8 @@ where
     where
         R: CryptoRng + RngCore + ?Sized,
     {
-        let mut rand_var_1 = Variable::default();
-        let mut rand_var_2 = Variable::default();
+        let mut rand_var_1 = self.zero_var();
+        let mut rand_var_2 = self.zero_var();
         // Blinding wires
         for _ in 0..2 {
             rand_var_1 = self.add_input(F::rand(rng));
--- a/plonk-core/src/constraint_system/composer.rs
+++ b/plonk-core/src/constraint_system/composer.rs
@@ -212,7 +212,7 @@ where
             composer.add_witness_to_circuit_description(F::zero());
 
         // Add dummy constraints
-        composer.add_dummy_constraints();
+        composer.add_blinding_factors();
 
         composer
     }
@@ -467,9 +467,9 @@ where
         f_x
     }
 
-    /// This function is used to add a blinding factor to the witness
-    /// polynomials. It essentially adds two dummy gates to the circuit
+    /// This function adds two dummy gates to the circuit
     /// description which are guaranteed to always satisfy the gate equation.
+    /// This function is only used in benchmarking
     pub fn add_dummy_constraints(&mut self) {
         // Add a dummy constraint so that we do not have zero polynomials
         self.q_m.push(F::from(1u64));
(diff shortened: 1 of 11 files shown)


plonk-core/src/constraint_system/composer.rs
```diff
@@ -526,6 +526,74 @@ where
         self.n += 1;
     }
 
+    /// This function is used to add a blinding factors to the witness
+    /// and permutation polynomials.
+    /// All gate selectors are turned off to guarantee the constraints
+    /// are still satisfied.
+    pub fn add_blinding_factors(&mut self) {
```
bhgomes: Can this take the RNG as a parameter?

```rust
pub fn add_blinding_factors<R>(&mut self, rng: &mut R)
where
    R: CryptoRng + RngCore + ?Sized;
```

plonk-core/src/constraint_system/variable.rs
```diff
@@ -10,7 +10,7 @@
 
 /// The value is a reference to the actual value that was added to the
 /// constraint system
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Default)]
```
bhgomes: I don't think it makes sense for the user to be able to construct a `Variable`. If we allow `Default`, we can't enforce this. What's the specific place we need `Default`?

bhgomes (commented):
Thanks for the implementation @davidnevadoc! Just some comments, while you're still in draft.

plonk-core/src/constraint_system/composer.rs
bhgomes: Also, since the gate selectors are off, we have to make sure this doesn't get disabled by the "empty selectors shouldn't contribute to the proof" mechanism. Perhaps we need to check that there is a non-trivial permutation on the specific gate as well.

plonk-core/src/constraint_system/variable.rs
davidnevadoc: I used it here: 
https://github.com/ZK-Garage/plonk/pull/118/files#diff-7580c65a30bf9cd380a7b494e25c0fd4b5760734af32c97a485d9e1d9b52b1f8R534-R535

but you are right. It can be changed for the 0 Variable for example.

plonk-core/src/constraint_system/composer.rs
davidnevadoc: I don't think that the unused selectors optimizations will interfere with the blinders.
The way I see it, those optimizations only affect:
 -  the quotient and lin polys, which are not affected by the blinders anyways 
 -  the proof and verifier keys which may omit some evaluations/commitments of unnecessary selectors, but everything witness realted stays unchanged.
 
 In regards to the non-trivial permutation I'm still working on guaranteeing that we wont get the trivial permutation in the whole circuit. (An attempt prevent this is the addition of the `zero_var()` in the 3rd blinder gate: https://github.com/ZK-Garage/plonk/pull/118/files#diff-7580c65a30bf9cd380a7b494e25c0fd4b5760734af32c97a485d9e1d9b52b1f8R592-R598).
 Is this what you mean by "non-trivial permutation on the specific gate"?

CPerezz (approved):
LGTM!!

Might be nice to get some benchmark results and publish them in the PR so that we can see if this causes significant perf regressions

plonk-core/src/constraint_system/composer.rs
```diff
@@ -526,6 +527,77 @@ where
...
 
+    /// This function is used to add a blinding factors to the witness
+    /// and permutation polynomials.
+    /// All gate selectors are turned off to guarantee the constraints
+    /// are still satisfied.
+    pub fn add_blinding_factors<R>(&mut self, rng: &mut R)
+    where
+        R: CryptoRng + RngCore + ?Sized,
+    {
+        let mut rand_var_1 = self.zero_var();
```
markulf, line 538: Whats the reason for moving `rand_var_1-2` out and not `rand_var_3-4`. Simplify or explain!

The reason is that these variables will be coppied into witness polynomials twice.

plonk-core/src/constraint_system/composer.rs
```diff
@@ -718,11 +790,10 @@ mod test {
         // NOTE: Circuit size is n+3 because
```
markulf: How is the 4 in the assert related to the 3 in n+3? Should latter be 4?

plonk-core/src/constraint_system/composer.rs
```diff
@@ -212,7 +213,7 @@ where
             composer.add_witness_to_circuit_description(F::zero());
 
         // Add dummy constraints
-        composer.add_dummy_constraints();
+        composer.add_blinding_factors(&mut rand::rngs::OsRng);
```
markulf, line 216: do we have general guidelines for generating randomness?

plonk-core/src/constraint_system/composer.rs
```diff
@@ -526,6 +527,77 @@ where
...
+        self.q_o.push(F::zero());
+        self.q_c.push(F::zero());
+        self.q_4.push(F::zero());
+        self.q_arith.push(F::zero());
+        self.q_range.push(F::zero());
+        self.q_logic.push(F::zero());
+        self.q_fixed_group_add.push(F::zero());
+        self.q_variable_group_add.push(F::zero());
+
+        self.perm.add_variables_to_map(
```
markulf, line 592: this call with the same variables is enough to randomize the permutation polynomial Z.

What is meaning of variables for which add_variables_to_map is not called?

plonk-core/src/constraint_system/helper.rs
```diff
@@ -46,12 +46,8 @@ where
...
 {
     // Common View
-    let universal_params = PC::setup(
-        2 * n + 6, // +1 per wire, +2 for the permutation poly
-        None,
-        &mut OsRng,
-    )
-    .map_err(to_pc_error::<F, PC>)?;
+    let universal_params =
+        PC::setup(2 * n, None, &mut OsRng).map_err(to_pc_error::<F, PC>)?;
```
markulf, line 50: What is n? circuit_size? How is it related to n+3/4 of `check_circuit_satisfied`?
Can't it be read off gadget?

Improve documentation of input arguments to `gadget_tester`.

plonk-core/src/permutation/mod.rs
```diff
@@ -823,7 +823,7 @@ mod test {
 
         let sigmas: Vec<Vec<F>> = cs
             .perm
-            .compute_sigma_permutations(7)
+            .compute_sigma_permutations(cs.circuit_size())
```
markulf, line 826: This is an improvement. But could we simplify this even further? E.g. why not have `cs.compute_sigmas()`?

Would need to understand the context better to recommend anything.

plonk-core/src/proof_system/quotient_poly.rs
```diff
@@ -52,54 +52,54 @@ where
...
-    z_eval_8n.push(z_eval_8n[4]);
-    z_eval_8n.push(z_eval_8n[5]);
-    z_eval_8n.push(z_eval_8n[6]);
-    z_eval_8n.push(z_eval_8n[7]);
+    let mut z_eval_4n = domain_4n.coset_fft(z_poly);
+    z_eval_4n.push(z_eval_4n[0]);
+    z_eval_4n.push(z_eval_4n[1]);
+    z_eval_4n.push(z_eval_4n[2]);
+    z_eval_4n.push(z_eval_4n[3]);
+    // z_eval_8n.push(z_eval_8n[4]);
```
markulf: clean up?

markulf (commented):
Some minor comments, LGTM otherwise.

plonk-core/src/constraint_system/composer.rs
davidnevadoc: `add_variables_to_map` is called for all variables every time they are used in a gate. This is used to create a map from which we extract the permutations that enforce the wire constraints of the circuit.  
So a variable for which this function is not called would be something like an input to a gate that is not used anywhere else in the circuit and is "disconnected". However the variable should be added anyways.

plonk-core/src/constraint_system/composer.rs
davidnevadoc: Yes, it should be a 4 :+1:

plonk-core/src/constraint_system/helper.rs
davidnevadoc: The input `n` here is an upper bound to the circuit size ( the number of constraints). We use `2*n` because we always have `2 *n > n.next_power_of_two()`.

 You're right, we should improve this doc.

plonk-core/src/constraint_system/composer.rs
davidnevadoc: Not that I know of TBH...
Any idea @CPerezz, @bhgomes ?

bhgomes: We use the cryptographic PRNGs provided by the `rand` library which use OS-provided entropy for seeding. I think we should eventually push the PRNG arguments to the caller, but at the moment this is what we're doing.

PR #120, opened 2022-02-18
Faster tests

### LINKED ISSUE #119: Slow tests
Running the full set of tests takes up an excessive amount of time, specially in not so powerful machines. This is makes development over all slower.
I think this problem can be mitigated easily with 2 small changes:
  - Reducing the SRS in `test_full` form `2^19` to `2^10` which is enough for the circuit.
  - Modifying the kzg tests name (adding a `_kzg` at the end) for easier filtering. This way we can run all the tests but just with kzg and just in one curve.

davidnevadoc:
Closes  #119 .

Further improvements could be made by generating the SRS once and using it for all tests that need it.


iquerejeta (approved):

LukePearson1 (approved):
LGTM!

PR #124, opened 2022-03-24
Plook up separate lookups

lopeetall:


plonk-core/src/proof_system/permutation.rs
```diff
@@ -53,13 +53,11 @@ where
...
 }
 
-//TODO Define structs to tidy up function inputs
 impl<F> ProverKey<F>
 where
     F: FftField,
 {
-    /// Computes permutation term of the quotient polynomial at the `i`th domain
-    /// point.
+    /// Computes the quotient polynomial at the `i`th domain point.
```
davidnevadoc: I think the previous comment was more accurate, since this function doesn't return evaluation of the full quotient polynomial, the gate constraint and lookup contributions are missing.

plonk-core/src/proof_system/permutation.rs
```diff
@@ -215,12 +151,7 @@ where
         (z_i - F::one()) * l1_alpha_sq
     }
 
-    // // L0(X) * (z2(X) − 1) * α_1^5
-    fn compute_lookup_term_check(&self, z_2_i: F, l1_alpha_five: F) -> F {
-        (z_2_i - F::one()) * l1_alpha_five
-    }
-
-    /// Computes the permutation term of the linearisation polynomial.
+    /// Computes the linearisation polynomial.
```
davidnevadoc: ditto

plonk-core/src/proof_system/widget/lookup.rs
```diff
@@ -81,3 +209,78 @@ where
...
+                + zeta * evaluations.wire_evals.b_eval
+                + zeta * zeta * evaluations.wire_evals.c_eval
+                + zeta * zeta * zeta * evaluations.wire_evals.d_eval;
+            f_0 * lookup_sep
+        };
+
+        scalars.push(f);
+        points.push(self.q_lookup.clone());
+
+        // (1 + δ) * (ε + f_bar) * (ε(1+δ) + t_bar + δ*tω_bar) *  lookup_sep^2
```
davidnevadoc, line 273: + L_1 * lookup_sep^3

davidnevadoc (changes requested):

plonk-core/src/proof_system/permutation.rs
lopeetall: Yeah I completely agree I don't know what I was thinking there

plonk-core/src/proof_system/widget/lookup.rs
lopeetall: line 267 has this computation

plonk-core/src/proof_system/widget/lookup.rs
```diff
@@ -210,71 +210,78 @@ where
...
-            let f_0 = evaluations.wire_evals.a_eval
-                - evaluations.lookup_evals.f_eval
-                + zeta * evaluations.wire_evals.b_eval
-                + zeta * zeta * evaluations.wire_evals.c_eval
-                + zeta * zeta * zeta * evaluations.wire_evals.d_eval;
-            f_0 * lookup_sep
-        };
+        // f =q_lookup * (lc([a_eval, b_eval, c_eval, d_eval] , zeta) - f_eval)
+        // lookup_sep is the eval of q_lookup and should be in the proof
+
```
lopeetall, line 257: I don't think that is true. q_lookup doesn't work as a selector on the lookup permutation terms which  need a separator

plonk-core/src/proof_system/widget/lookup.rs
davidnevadoc: but we should add it in the comment too!

lopeetall: Can you explain this comment? I'm not sure what you are trying to point out.

lopeetall: My apologies I just saw your reply to this, I think it was on a different commit.

plonk-core/src/proof_system/preprocess.rs
```diff
@@ -297,6 +286,14 @@ where
             adicity:
                 <<F as FftField>::FftParams as ark_ff::FftParameters>::TWO_ADICITY,
         })?;
+
+        let preprocessed_table = PreprocessedLookupTable::<F, PC>::preprocess(
+            &self.lookup_table,
+            commit_key,
+            domain.size().try_into().unwrap(),
```
davidnevadoc: We can replace with simpler:
``` rust
domain.size() as u32,
```
As a side note, I believe we are using `u32` and `u64` for n and circuit size. It would be good to use only on type throughout all the code.

davidnevadoc (approved):
Just a small nit, but ready to merge!
Great work :)

PR #125, opened 2022-03-26
Plookup passing

lopeetall:
Plookup is fully integrated in this branch and all tests are now passing.

The version of Plookup used in this branch is based closely on the version in the Plonkup paper which was easiest to reference and debug. We can change the lookup argument to include new techniques from this point.

Some attempt was made to rename variables to make the debugging process easier, but the changes weren't made everywhere so I'm afraid variable names are still a bit of a mess.

Next steps:
- table preprocessing ought to be moved out of shared preprocessing so it can occur separately
- still plenty of clean up, documentation, and optimization to be done


plonk-core/src/proof_system/widget/lookup.rs
```diff
@@ -117,38 +172,94 @@ where
...
-            -z2_i_next * c_1 * c_2 * lookup_sep_sq
+            -z2_i_next * c_0 * c_1 * lookup_sep_sq
         };
 
         let d = { (z2_i - F::one()) * l1_i * lookup_sep_cu };
 
         a + b + c + d
     }
+    /// Compute evals of lookup portion of quotient polynomial
+    pub fn debug_compute_quotient_i(
```
davidnevadoc: should this function replace `compute_quotient_i`?

plonk-core/src/proof_system/widget/lookup.rs
```diff
@@ -93,6 +93,61 @@ where
             .collect())
     }
 
+    /// Compute lookup portion of quotient polynomial
+    pub fn debug_compute_lookup_quotient_term(
```
davidnevadoc: should this replace `compute_lookup_quotient_term`?

davidnevadoc (approved):
We still need to do a bit of work, mainly what you pointed out in the next steps of the PR.  The rest LGTM. Great work! :)

plonk-core/src/lookup/lookup_table.rs
```diff
@@ -34,7 +34,7 @@ where
     }
 
     /// Returns the length of the `LookupTable` vector.
-    fn len(&self) -> usize {
+    fn _len(&self) -> usize {
```
LukePearson1, line 37: Why has this function changed?

LukePearson1 (approved):
Left a small question. Other than that, looks good to me.

plonk-core/src/lookup/lookup_table.rs
lopeetall: The CI was complaining about it being unused. I didn't see an obvious place where it should actually *be* used but didn't want to remove it, in case I missed something, so I gave it the underscore tag.

PR #147, opened 2022-05-27
Update `PI` size in `check_circuit_satisfied`

### LINKED ISSUE #146: Bug report: check_circuit_satisfied doesn't work
panicked at here: index out of bounds
https://github.com/ZK-Garage/plonk/blob/27733c60fbe4e32c21e88a344d450d29e977a4bf/plonk-core/src/constraint_system/composer.rs#L685

`pi_vec.len()` is less than `self.n`, I guess that`self.public_inputs.n` is forgotten to update somewhere.
The previous version is fine because pi_vec uses `self. n` as length.

davidnevadoc:
Adds an extra `update_size()` to make sure PI has the correct length.
Close #146

### DIFF
--- a/plonk-core/Cargo.toml
+++ b/plonk-core/Cargo.toml
@@ -45,7 +45,7 @@ std = [
 trace = []
 
 # Tracing Printing Features
-trace-print = ["trace"]
+trace-print = ["trace", "std"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


CPerezz (approved):
LGTM!

weikengchen (approved):

PR #155, opened 2022-06-17
Add test for permutation argument constants.

### LINKED ISSUE #154: Testing coset constants
The constants we are currently using for generating `H` cosets for the permutation argument are correct for the fields currently used (`Bls` scalar fields). However we should have a test that ensures these constants are valid. Specially considering the addition of more wires (and therefore more constants) and other supported fields.

davidnevadoc:
Close #154

### DIFF
--- a/plonk-core/src/permutation/constants.rs
+++ b/plonk-core/src/permutation/constants.rs
@@ -2,11 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //
-// Copyright (c) DUSK NETWORK. All rights reserved.
+// Copyright (c) ZK-GARAGE. All rights reserved.
 
-//! Constants used in the permutation argument to ensure that the wire subsets
-//! are disjoint.
-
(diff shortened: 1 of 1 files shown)


CPerezz (approved):
How much time doesnthis take to run?? Wondering if we should ignore or not.

Aside from that, great work mate!

akinovak:
Wow looks so great and clean! Should we also put a hackmd link on top of the file?

@CPerezz for not I guess it might be ignored, but I think we need to include it as soon as we switch to arbitrary number of wires.

davidnevadoc:
This runs pretty quick. Takes `O(k^2)` multiplications with `k` being the number of wires which its usually not too large.
I think right now having the constants hardcoded this is fine to have as a test. If at some point we generate them in some other way this may be part of the generation algorithm.

weikengchen:
Why is `prev_cts` never updated?

davidnevadoc:
> Why is `prev_cts` never updated?

I forgot to update it at the end of the loop. Well spotted! :+1:

weikengchen (approved):

weikengchen:
I feel it is ready to merge.

PR #159, opened 2022-07-21
Add a new example

bwuzhang:
As someone new to this PLONK repo, I found it would be nice to have more examples besides the only `simple_circuit.rs`.

This example demonstrates the usage of gate chaining and the following gates: 

- add
- mul
- is_zero_with_output
-  xor_gate

### DIFF
--- a/examples/example_poly.rs
+++ b/examples/example_poly.rs
@@ -1,0 +1,166 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+//
+// Copyright (c) ZK-GARAGE. All rights reserved.
+
+//! PLONK Example
+
+use ark_bls12_381::{Bls12_381, Fr as BlsScalar};
+use ark_ec::{TEModelParameters};
+use ark_ed_on_bls12_381::{
+    EdwardsParameters as JubJubParameters, Fr as JubJubScalar,
+};
+use ark_ff::PrimeField;
+use ark_poly::polynomial::univariate::DensePolynomial;
+use ark_poly_commit::{sonic_pc::SonicKZG10, PolynomialCommitment};
+use plonk::error::to_pc_error;
+use plonk_core::circuit::{verify_proof, Circuit};
+use plonk_core::constraint_system::StandardComposer;
+use plonk_core::error::Error;
+use plonk_core::prelude::*;
+use rand_core::OsRng;
+
+fn main() -> Result<(), Error> {
+    // Implements a circuit that defines the following function:
+    // -------------------
+    // def f(x, y, z):
+    //   if x == 1:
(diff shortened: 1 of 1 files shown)


examples/example_poly.rs
```diff
@@ -0,0 +1,164 @@
...
+            });
+
+            // Define -z
+            let minus_z = composer.arithmetic_gate(|gate| {
+                gate.witness(z, minus_1, None).mul(F::one())
+            });
+
+            // Define 2y-z
+            let two_y_minus_z = composer.arithmetic_gate(|gate| {
+                gate.witness(two_y, minus_z, None).add(F::one(), F::one())
```
davidnevadoc: Since `2` is a constant and the subsctraction is just an addition with an implied `-1` constant, we can eliminate the inputs: `one`, `two` and `minus_1` as well as the previous gates that compute `two_y` and `minus_z`. We can fit all the computation in just one gate:
``` rust
let two_y_minus_z = composer.arithmetic_gate(|gate| {
    gate.witness(y, z, None).add(F::from(2u64), -F::one())
});
```

examples/example_poly.rs
```diff
@@ -0,0 +1,164 @@
...
+            });
+
+            // Define y*z
+            let y_times_z = composer
+                .arithmetic_gate(|gate| gate.witness(y, z, None).mul(F::one()));
+
+            // Define x-1
+            let x_minus_1 = composer.arithmetic_gate(|gate| {
```
davidnevadoc, line 79: In a similar way here we can have:
``` rust
let x_minus_1 = composer.arithmetic_gate(|gate| {
    gate.witness(x, zero, None)
        .add(F::one(), F::zero())
        .constant(-F::one())
});
```
(having previously declared `let zero = composer.zero_var()`)

examples/example_poly.rs
```diff
@@ -0,0 +1,164 @@
...
+                gate.witness(x, minus_1, None).add(F::one(), F::one())
+            });
+
+            // Define x-1==0
+            let x_bool = composer.is_zero_with_output(x_minus_1);
+
+            // Define x bool negate
+            let x_bool_negate = composer.xor_gate(x_bool, one, 10);
```
davidnevadoc, line 91: This works fine, but we can go for a simpler and cheaper way. Since `x_bool` is guaranteed to be either `0` or `1` we can do 
` x_bool_negate = 1 - x_bool` like this:
``` rust
let x_bool_negate = composer.arithmetic_gate(|gate| {
    gate.witness(zero, x_bool, None)
        .add(F::zero(), -F::one())
        .constant(F::one())
});
```
However, this is ok for the purpose of showcasing `xor_gate` in an example :)

davidnevadoc (commented):
Thanks a lot for the contribution!! :)

I added a couple of comments regarding how the handling of constants can be improved. 
Other than that, great work! I think the example does a good job of showing how the library can be used :+1:

bwuzhang:
@davidnevadoc thanks for the review! Comments addressed in the new commit.

davidnevadoc (approved):
LGTM! :+1:

LukePearson1 (approved):
Thanks for the PR. All looks good to me. 

Will merge in.

PR #160, opened 2022-07-29
Addition circuit requires the A coefficient

simonmasson:
Custom addition gate is implemented for twisted Edwards curves with coefficient `a = -1`. We modify it so that it works for curves that do not have this property (Bandersnatch, or new BLS12 curves).


simonmasson:
@lopeetall @LukePearson1 can you review this? I don't know how to add reviewers...

davidnevadoc (approved):
Thanks for the PR! I think this is a good change to have.
LGTM :+1:

LukePearson1 (approved):
LGTM!

PR #161, opened 2022-07-29
Example fix

simonmasson:
In the example, the `range_gate` is used in the wrong way: it checks that the `witness < 2^num_bits`. In the example, we actually check that `5 < 2^(2^5)` which is confusing.


simonmasson:
@lopeetall @LukePearson1 can you review this PR? I don't know how to add reviewers...

davidnevadoc (approved):
Well spotted! Thanks for the fix!

LukePearson1 (approved):
LGTM!

